PR #4, opened 2019-07-16
Added Code for Problem1.java -- Please Review

lavanganji:
not working with another input [1,1]

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,36 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n) + map space. ie.e O(n) 
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : No , but the didnt get the approach.
+
+
+// Your code here along with comments explaining your approach
+        // thinking about the brut force.
+        // binary search for missing element if the array is in sorted , i.e. O(log n)
+        // Bitwise XOR for missing element if the array is in not sorted  , i.e. dont know.
+        
+
+
+class Solution {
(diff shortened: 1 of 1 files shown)


Problem1.java
```diff
@@ -0,0 +1,36 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n) + map space. ie.e O(n) 
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : No , but the didnt get the approach.
+
+
+// Your code here along with comments explaining your approach
+        // thinking about the brut force.
+        // binary search for missing element if the array is in sorted , i.e. O(log n)
```
ashmeetdeol19: That wont be efficient if more than one elements are missing.

Problem2.java
```diff
@@ -0,0 +1,54 @@
...
+
+public static void main(String[] args){
+int[] arr = {4,3,2,7,8,2,3,1};
+Pair p = getMinMax(arr);
+System.out.print("Min : ",p.minArr);
+System.out.print("Max : ",p.maxArr);
+}
+}
+
+
```
ashmeetdeol19: What are the no. of comparisons in your solution ?

Problem1.java
lavanganji: yes, just implemented it and it takes more time. also included the better solution.

Problem2.java
lavanganji: it should be 2(n-1)

PR #7, opened 2019-07-17
Completed Problem 1

gauravj09:


problem1.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+            }
+
+            for(int i = 0; i<nums.size(); i++)
+                if(nums[i] > 0) missingValues.push_back(i+1);
+            
+
+            return missingValues;
+        }
+};;
```
ashmeetdeol19: Thats a great solution with good explanation. 👍

PR #9, opened 2019-07-17
Completed Array 2 - Problem 1 and Problem2 

bharatreddypothireddy:


Problem2.java
```diff
@@ -0,0 +1,31 @@
...
+        }
+        return new int[]{min,max};
+    }
+    public static void main(String[] args) {
+        int [] nums = {5,4,3,2,1};
+        int [] result = minMax(nums);
+        System.out.println("Min : "+result[0]);
+        System.out.println("Max : "+result[1]);
+    }
+}
```
ashmeetdeol19: Well Done but That's not the required solution.
We are supposed to reduce the no. of comparisons.

Problem1.java
```diff
@@ -0,0 +1,30 @@
...
+                list.add(i+1);
+        }
+
+        return list;
+    }
+    public static void main(String[] args) {
+        int [] nums = {4,3,2,7,8,2,3,1};
+        System.out.println(findDisappearedNumbers(nums));
+    }
+}
```
ashmeetdeol19: Thats a good one 👍

PR #26, opened 2019-09-03
Problem 1 (LC verified )and 2(not on leetcode) for Array-2 section.

manishjangid:


Problem2.c
```diff
@@ -0,0 +1,79 @@
...
+    }
+    /* Initialize min and max */
+    min = nums[0];
+    max = nums[1];
+
+    /* Go over the numbers from 3rd element onwards and based on condition update min and max.*/
+    for (idx = 2; idx < numsSize; idx++) {
+        if (min > nums[idx]) {
+            min = nums[idx];
+        } else if (max < nums[idx]) {
```
kaurramanpreet: why have you added else if condition?
is else required?

manishjangid: Hi Raman, I have added else if condition as I was not able to find any case where a min is updated and then max will also get updated in the same condition. Let me know if I am wrong here.
Also I just realised that I set min max for the first 2 elements without even comparing them. I will fix it too.

Problem3.c
```diff
@@ -0,0 +1,218 @@
...
+                          [0,0,1],     temp_array[1] = 1,0,1
+                          [1,1,1],    now original[0] overwritten by temp_array[0] and temp_array[0] gets new result
+                                       ie 0,1,1
+                          [0,0,0]  now original[1] overwritten by temp_array[1] and temp_array[1] gets new result
+                                       ie 0,1,0
+                        ]
+             * When row becomes 2 then we start copying the data in the manner shown above
+             *
+             *
+             */
```
kaurramanpreet: amazing explanation!!

PR #38, opened 2019-09-05
Previous Problems

KaranDar:
@kaurramanpreet @jaspindersingh83


kaurramanpreet:
try to submit the prs on time and also before the sessions
this will help you in learning more in the class!!

kaurramanpreet:
your all the codes are correct but the best way to improve is when you start thinking of an alternative solution when you are able to come up with one of the solutions for the problem!!

KaranDar:
Thanks @kaurramanpreet I will try to be on time :)

PR #49, opened 2019-12-03
Completed all the problems - Please Review

devashishnyati:
Will try gameoflife again after the class

### DIFF
--- a/getMinMax.py
+++ b/getMinMax.py
@@ -1,0 +1,35 @@
+# Time Complexity : Brute Force: O(n)
+# Space Complexity : Brute Force: O(n)
+# Did this code successfully run on Leetcode : Yes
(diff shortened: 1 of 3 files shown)


getMinMax.py
```diff
@@ -0,0 +1,35 @@
+# Time Complexity : Brute Force: O(n)
+# Space Complexity : Brute Force: O(n)
+# Did this code successfully run on Leetcode : Yes
```
neha79091: why is your space complexity O(n)

jaspindersingh83: @neha79091 He is right the time complexity is O(n) only. The number of comparisons though is 2n.

jaspindersingh83: Yup we are not using any extra DS here Devasis

PR #50, opened 2019-12-03
Completed Array-2

RishikaMachina:


Problem_2.py
```diff
@@ -0,0 +1,26 @@
+# Space complexity:  O(1)
+# Runtime complexity:  O(n)
+# number of comparisions will decrease by 2 as grouping of elements is done.
+
+# Didn't find the question on Leetcode, Can I know the question number? I will update the code after running on leetcode
+
```
neha79091: this problem is not on leetcode.

RishikaMachina: Okay, Is this a good solution then?

RishikaMachina:
Okay, will include from next submissions.

RishikaMachina:
Included the description. Please review.

PR #52, opened 2019-12-03
Completed 1 problem. Will complte other 2 after class

Ajayvikram10:


DisappearedNumber.java
```diff
@@ -0,0 +1,32 @@
+import java.util.*;
+
+public class DisappearedNumber {
+
+	//	Time:  O(m*n) 
+	//	Space O(m*n)
+	//	LeetCode: yes
```
neha79091: Please give information about m and n

PR #66, opened 2020-01-08
Solved all three problems. Please review.

harimanasa:


rajatkumar01:
Good work for all the problems

PR #82, opened 2020-02-18
Create GameOfLife.py

megrao:
Still modifying the code


Lohith-Aswathappa:
Nice try, Megha!

Try to optimize the solution after class

PR #83, opened 2020-02-18
Array-2 before class @nikrasiya

sowmyadvn:


BhushanVasisht:
Good Job Sowmya. Try the follow up question of minimizing number of comparisons too

BhushanVasisht:
Good Job Sowmya. Try the follow up question of minimizing number of comparisons too

sowmyadvn:
Tried that approach too. Please review. Thanks!

super30admin:
Yup correct

PR #88, opened 2020-02-18
Please review

PratikshaPP:
Tried Game of Life but i was doing it wrong since i was updating the board and reusing the previous values.


Lohith-Aswathappa:
Nice try! Prathiksha

try to solve the other problems as well after the class

PR #107, opened 2020-03-31
Array-2

AMAbhi:
Please Review Problem-1 and Problem-2
I will submit Problem 3 after discussion in the class.

### DIFF
--- a/Problem2.py
+++ b/Problem2.py
@@ -1,0 +1,41 @@
+#Time Complexity-O(n)
+#Space Complexity-O(1)
+
+class Solution:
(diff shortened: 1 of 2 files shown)


Problem2.py
```diff
@@ -0,0 +1,41 @@
+#Time Complexity-O(n)
+#Space Complexity-O(1)
```
devashishnyati, line 2: You are using extra space of a set?

devashishnyati:
Good job Abhishek!
Optimize problem 2.
Please submit other problem also.

PR #110, opened 2020-04-01
adding the codes for array2

shahakshay11:


game_of_life.py
```diff
@@ -0,0 +1,72 @@
+"""
+// Time Complexity : O(m*n)
+// Space Complexity : O(m*n)
```
devashishnyati, line 3: Where are you using this extra space?

devashishnyati:
Good job Akshay!

game_of_life.py
shahakshay11: So I am updating the changed state of the board and adding to results array to tackle the problem of altering state for other cells in the array. So worst case it'll be m*n
But yeah, now with alias method where we can assign a dummy var to live and dead state discussed today, I'll update this PR. Thanks.

PR #114, opened 2020-04-01
problem1-2 completed

schakrad:


Arrays-2-problem-1.txt
```diff
@@ -0,0 +1,21 @@
+TC:O(N)
+SC:O(N)
```
devashishnyati, line 2: Try to optimize it on space.

devashishnyati:
Good job Sindhu.
Try to optimize problem 1 in space.
Submit other problem also.
Instead of submitting .txt files, you can submit .java files also.

PR #118, opened 2020-04-01
Completed Array 2

AroopN:


devashishnyati:
Good job Aroop!
Submit the other problem as well.

PR #124, opened 2020-05-05
Problem30 - Please review

KGurpreet:


Problem30 - Please review
```diff
@@ -0,0 +1,25 @@
+// Time Complexity :O(n)
```
super30admin, line 1: This seems to be O(n2) solution. Can you discuss your solution in the class today please ?

super30admin:
This seems to be O(n2) solution. Can you discuss your solution in the class today please ?

Problem30 - Please review
KGurpreet: I will re-submit this PR. Will be working on it, have some confusions.

PR #125, opened 2020-05-05
Working solution for the 3 problems

domzhaomathematics:


sowmyadvn:
Good job Dominic! All the solutions are correct and well done with the explanation.

sowmyadvn:
Good job Dominic! All the solutions are correct and well done with the explanation.

PR #132, opened 2020-05-06
Completed Sprint-Array-2

Akshay-Ravi:


find_min_max.cpp
```diff
@@ -0,0 +1,38 @@
+//Time complexity - 3 comparisions are being done here within O(n)
```
super30admin, line 1: 3 comparisons are within a pair. 3 comparisions are being done here within O(n/2)

super30admin:
Good Job with 2 problems.

PR #136, opened 2020-05-07
Completed Design for Question 1 and 2

JacobStephen9999:


super30admin:
Solutions are correct. Name your git commits detailed and name them better

PR #138, opened 2020-05-07
Completed PRs-Please review

sbsreedh:


MinMax.py
```diff
@@ -0,0 +1,35 @@
+# Time Complexity :O(n)
+# Space Complexity :O(log n)-For recursion call stack
+
+#If arr size is 1: return the first element as min and max
+#If the array size is 2: Compare between two numbers and return min and max
+#Else recursively store and calculate min and max of the left and right array.
```
super30admin, line 11: Not a good approach to call recursively on left side of array and right side of array. Its as good as linear search. The comparisons in given solution are still O(2n)

super30admin:
Thumb rule. Never apply binary search on left side of array and right side of array. Do you know why?

PR #148, opened 2020-06-08
Completed : Please Review

ashiranka:


AbhiniveshP:
Nice work !!!
Please try without extra space for find disappeared numbers question and also include the solution for other question as well.

AbhiniveshP:
Nice work !!!
Please try without extra space for find disappeared numbers question and also include the solution for other question as well.

PR #151, opened 2020-06-09
Array - 2

Sumesh-H:


MissingNum.java
```diff
@@ -0,0 +1,44 @@
...
+            if(!numbers.contains(i)){
+                missingNumbers.add(i);
+            }
+        }
+        return missingNumbers;
+    }
+}
+
+// with no extra space O(1)
+
```
Sumesh-H, line 27: Here the approach is, we iterate through the input array and mark elements as negative using nums[nums[i]-1] = -nums[nums[i]-1]. In this way, all the numbers that we have visited will be marked as negative. In the second iteration, if a value is not marked as negative, it implies we have never visited that index before, so just add it to the return list.

AroopN:
Good work, Sumesh.

PR #155, opened 2020-06-09
MissingNumbersOfArray_GameOfLife

tramatejaswini:


MissingNumbersArray.py
```diff
@@ -0,0 +1,24 @@
+  ---------------------- Missing numbers in the array------------------------------------------------   
+# Time Complexity : O(N) N is the length of the nums.
+# Space Complexity : O(1) as I am returning the same array used for exploring numbers
```
AbhiniveshP, line 3: There is a usage of set(nums) and Space complexity would not be O(1)

AbhiniveshP:
Nice work !!!

PR #163, opened 2020-06-09
Completed Array-2-Please Check

DineshNadar95:


MinMaxArray.java
```diff
@@ -0,0 +1,62 @@
+// Time Complexity :O(n/2) I am a bit confused can you xplain on this one.
```
AbhiniveshP, line 1: The Time Complexity would be O(3n/2) but we ignore constants and so it would be O(n).
Here, instead of traversing the array 2n times, we traverse it only 3n/2 times as we pair 2 elements each time. This reduces traversing the array 2 times to 3n/2 times. I hope you understand !!!

AbhiniveshP:
Nice work Dinesh !!!
Reply back here if you still didn't understand the time complexity for min_max of array.

PR #176, opened 2020-06-10
Array-2 done

Akash-Akki:


AbhiniveshP:
Nice work on both these questions.
Also, include the solution for min_max problem.

AbhiniveshP:
Nice work on both these questions.
Also, include the solution for min_max problem.

PR #181, opened 2020-06-11
Completed Array-2: Please review

shantanu609:


AbhiniveshP:
Nice work on all 3 questions Shantanu !!!
You could've mentioned the significance of min_max problem to be reducing the number of operations from around 2n to around 3n/2.

AbhiniveshP:
Nice work on all 3 questions Shantanu !!!
You could've mentioned the significance of min_max problem to be reducing the number of operations from around 2n to around 3n/2.

PR #187, opened 2020-06-15
Array 2 Problems: Please review it

nareshkumar-ss:


AbhiniveshP:
The number of comparisons for min_max problem should be less than 2n and the optimal number of comparisons end up being around the order of (3n/2). Please try implementing that solution.

AbhiniveshP:
The number of comparisons for min_max problem should be less than 2n and the optimal number of comparisons end up being around the order of (3n/2). Please try implementing that solution.

PR #202, opened 2020-07-23
Completed. Please review it

sneharnair:


Problem-3_Game_of_life.py
```diff
@@ -0,0 +1,101 @@
+# APPROACH 1: USING EXTRA SPACE
+# Time Complexity : O(m*n), m: number of rows of board, n: number of columns of board
+# Space Complexity : O(m*n), the space of result array
```
samarthjj: The SC should be O(1), result array is not auxiliary

Problem-3_Game_of_life.py
```diff
@@ -0,0 +1,101 @@
...
+                    result[row][column] = board[row][column]
+                        
+        return result
+        
+        
+        
+
+# APPROACH 2: MODIFY IN-PLACE 
+# Time Complexity :  O(m*n), m: number of rows of board, n: number of columns of board
+# Space Complexity : O(m*n), the space of result array
```
samarthjj: SC should be O(1), board modified in place

samarthjj:
Nice work Sneha! Please review the inline comments.

samarthjj:
Nice work Sneha! Please review the inline comments.

PR #215, opened 2020-07-28
Completed Array 2

jvenkat:


gameoflife.py
```diff
@@ -0,0 +1,42 @@
+#Time Complexity O(n^2)
```
Sumesh-H, line 1: How? where n is?

minandmax.py
```diff
@@ -0,0 +1,13 @@
+class Solution:
```
Sumesh-H, line 1: TC & SC?

Sumesh-H (commented):
Good work, Venkat. Kindly check inline comments.

Sumesh-H:
Good work, Venkat. Kindly check inline comments.

PR #259, opened 2020-09-29
Problem 1 and 2

DhruvJawalkar:


DissapearedNumsInArray.java
```diff
@@ -0,0 +1,60 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : could get with O(n) extra space but didn't get the intuition for O(1)
```
kannan-mrugapphan, line 4: The Constant Space solution is based on the constraints that array has only numbers from 1 to n. This condition leads us to mark the respective indices as negative (if the number is present). Thus the algorithm is based on constraints.

MinMaxOfArray.java
```diff
@@ -0,0 +1,127 @@
+// Time Complexity : O(log n), for diving the problem in 2 equal halves
```
kannan-mrugapphan, line 1: Please check on the complexities. I feel that its still O(n) - 2T(n / 2) + 1

MinMaxOfArray.java
```diff
@@ -0,0 +1,127 @@
...
+// Space Complexity : O(log n), for the stack space
+// I'm not sure about how to calculate total comparisions in this method
```
kannan-mrugapphan, line 3: In the recursive case, you are doing 3 comparisons in each call, so it approximates to 3(n / 2) - number of elements is halved at every call.

kannan-mrugapphan (commented):
Good Job. Please try the other problem as well.

kannan-mrugapphan:
Good Job. Please try the other problem as well.

PR #261, opened 2020-09-30
array-2 done 

Vjais:


ramya-sruthi-d (commented):
Good work. But why is space complexity O(mn) in the game of life problem?

ramya-sruthi-d:
Good work. But why is space complexity O(mn) in the game of life problem?

PR #267, opened 2020-09-30
Array-2

HarshaVardhan-Kaki:


ramya-sruthi-d (commented):
Good work. For the game of life problem use directions array instead of if statements.

ramya-sruthi-d:
Good work. For the game of life problem use directions array instead of if statements.

PR #269, opened 2020-10-01
Arrays-2 solved first problem

Jaisu-1:


ramya-sruthi-d (commented):
Nice work. Include your approach also in the code.

ramya-sruthi-d:
Nice work. Include your approach also in the code.

PR #300, opened 2020-10-29
Array2-submission

sujana99:
partial submission. second problem works for new result matrix. working on in matrix type submission.


samarthjj:
Good work Suajana, Please complete the other problems as well.

samarthjj:
Good work Suajana, Please complete the other problems as well.

PR #338, opened 2021-01-06
Self attempt pls review

nilesh-525:
I coulnd't come up with approch for minimising comparisons for finding min,mix.

Problem 3 is still pending.

### DIFF
--- a/FindMinMaxinArr.java
+++ b/FindMinMaxinArr.java
@@ -1,0 +1,38 @@
+import java.util.*;
+
+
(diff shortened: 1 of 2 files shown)


khushal1198 (commented):
Good work Nilesh. Now that you're done with the third problem in class, please submit it soon.

khushal1198:
Good work Nilesh. Now that you're done with the third problem in class, please submit it soon.

PR #345, opened 2021-01-07
Completed Array-2 - Please review

pranati05:


ramya-sruthi-d (commented):
Good job, you can optimize the sc for the first problem try that as well.

ramya-sruthi-d:
Good job, you can optimize the sc for the first problem try that as well.

PR #346, opened 2021-01-07
Completed Problem 1

shamali21:


keshav-Gupta-IS (commented):
Good work with the solution. Complete the other problem as well and write explanation of your approach.

keshav-Gupta-IS:
Good work with the solution. Complete the other problem as well and write explanation of your approach.

PR #383, opened 2021-02-02
Create find all numbers disappeared

srivastavavertika201:


find all numbers dissappeared
```diff
@@ -0,0 +1,19 @@
+class Solution {
+    public List<Integer> findDisappearedNumbers(int[] nums) {
+        List<Integer> list = new ArrayList<Integer>();
```
yahska-ednihs, line 3: Good Job! It would be great if you can add few comment in your code explaining your approach.

Game_Of_Life
```diff
@@ -51,4 +51,6 @@ class Solution {
 }
 
 
-// Need to work on it!
+//time complexity : O(n X m) = O(n)
+//space complexity : O(1)
```
yahska-ednihs, line 55: Good Job! You can add a few comments when your dry run on your code or before which will help reduce the streamline your understanding of your code.

PR #430, opened 2021-03-06
Array-2 please review!

ronak-prog:


khushal1198 (commented):
Good work Ronak!

khushal1198:
Good work Ronak!

que-3.java
```diff
@@ -0,0 +1,59 @@
...
+                    minmax.min = arr[i];
+                }  }
+        }
+    return minmax;
+    }
+    public static void main(String args[]) {
+        int arr[] = {1000, 11, 445, 1, 330};
+        int arr_size = 5;
+        Pair minmax = getMinMax(arr, arr_size);
+        System.out.printf("\nMinimum element is %d", minmax.min);
```
aidanBurton, line 55: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

PR #487, opened 2021-06-08
Completed Array-2 - v1 - Please Review

isdarwhekar:


find_disappeared_numbers.cpp
```diff
@@ -0,0 +1,33 @@
+// Time Complexity : O(n)
+// Space Complexity :  O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this - Couldn't find the O(1) space solution
```
kannan-mrugapphan, line 4: Please try to use the fact that the array has numbers only between 1 and n to optimize on space.

kannan-mrugapphan (commented):
Good Job. For the second problem try optimizing the code a bit. There is a lot of if..else blocks reducing the readblity

kannan-mrugapphan:
Good Job. For the second problem try optimizing the code a bit. There is a lot of if..else blocks reducing the readblity

PR #493, opened 2021-06-09
Create problem1.cpp

pratikshakapse5:


sridhar-anish (commented):
Nice try make sure to try out the other problems as well

sridhar-anish:
Nice try make sure to try out the other problems as well

PR #541, opened 2021-06-15
Solved the 3

hharshitarora:


Rooppesh:
Good Job! Add the TC and SC!

Rooppesh:
Good Job! Add the TC and SC!

PR #617, opened 2021-09-01
Completed Sprint Array-2- Please Review

Yatha3011:


Pinak08 (commented):
SC for game of life problem is not m*n in your approach and also it can be done in O(1) SC

Pinak08:
SC for game of life problem is not m*n in your approach and also it can be done in O(1) SC

PR #652, opened 2021-10-06
completed array-2 please review

psnehas:


sanjayvarma199 (commented):
Good Job!! Submit other two problems as well

sanjayvarma199:
Good Job!! Submit other two problems as well

PR #888, opened 2022-06-08
"Completed Array-2"

ishankaggarwal:


Maximum and minimum of an array using minimum number of comparisons.cpp
```diff
@@ -0,0 +1,73 @@
+//Time Complexity- O(n)
+//Space Complexity- O(logn) //Stack space
```
Sumesh-H, line 2: Which Stack Space are you referring to?

Sumesh-H (commented):
Good Job, please address the inline comment.

Sumesh-H:
Good Job, please address the inline comment.

PR #889, opened 2022-06-08
Completed Sprint Array-2

daman1321:


Sakshi-2103 (commented):
Good work
Attempt other problems as well

Sakshi-2103:
Good work
Attempt other problems as well

PR #907, opened 2022-06-08
Solution for game of life & disappeared numbers

gauripas57:


niharikagupta2021 (commented):
Good Job Gauri! pl add sc and tc for both problems

niharikagupta2021:
Good Job Gauri! pl add sc and tc for both problems

PR #934, opened 2022-06-12
Completed. Please review.

Thirunaa:


Game of Life.py
```diff
@@ -0,0 +1,31 @@
+# TC : O(n)
```
ishika2741, line 1: Kindly check the complexity.

ishika2741 (commented):
Good work! Kindly submit third problem as well . Kindly explain your approach using comments.

ishika2741:
Good work! Kindly submit third problem as well . Kindly explain your approach using comments.

Game of Life.py
Thirunaa: Here by n I meant number of elements in the board. Its supposed to be O(m*n)  in actual terms.

PR #1046, opened 2022-09-14
Completed problems for array2

ysoni0303:


pvk3081 (commented):
Good Job Yash

pvk3081:
Good Job Yash

PR #1094, opened 2022-10-20
Completed - Array 2 - Please Review

Tanaya-Mahakalkar:


Exercise_2.py
```diff
@@ -0,0 +1,44 @@
+class Solution:
```
daredevilteja, line 10: Please mention TC and SC

daredevilteja:
Good Work! Keep it up

Exercise_2.py
```diff
@@ -0,0 +1,53 @@
+'''
+Time Complexity : O(n)
```
daredevilteja, line 2: TC is O(m*n)

daredevilteja:
Good Work! Keep it up

PR #1175, opened 2023-02-09
Completing Array-2 - Please review. 

Design55:


Exercise_3.py
```diff
@@ -0,0 +1,54 @@
+#Time Complexity : O(M*N), N being the length of the array and M being the width. 
+#Space Complexity : O(M*N), N being the length of the array and M being the width.
```
daredevilteja: SC is O(1) because no extra space is used here.

daredevilteja (changes requested):

daredevilteja:
Good Work! Keep it up.

Exercise_3.py
Design55: That is true.

Design55: Adjusted.

PR #1253, opened 2023-06-19
Array 2

salonimodi:


gameOfLife.py
```diff
@@ -0,0 +1,40 @@
+# Space Complexity: O(1) 
+# Time Complexity:  O(n)
```
ishika2741, line 2: It will be m*n

ishika2741 (commented):
Good work ! Keep it up.

ishika2741:
Good work ! Keep it up.

PR #1292, opened 2023-08-02
Completed Array-2

sana-h-pathan:
https://www.geeksforgeeks.org/maximum-and-minimum-in-an-array/
 https://leetcode.com/problems/game-of-life/
https://leetcode.com/problems/find-all-numbers-disappeared-in-an-array/

### DIFF
--- a/MaxMinInArray.java
+++ b/MaxMinInArray.java
@@ -1,0 +1,40 @@
+import java.util.ArrayList;
+import java.util.List;
+
+//TC=O(n)
+//SC=O(1)
+public class MaxMinInArray {
+
+    public List<Integer> findMaxMin(int[] nums){
+        List<Integer> result=new ArrayList<>();
+        int n=nums.length;
+        int maxValue=nums[0];
+        int minValue=nums[0];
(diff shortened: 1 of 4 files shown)


MaxMinInArray.java
```diff
@@ -0,0 +1,40 @@
+import java.util.ArrayList;
+import java.util.List;
+
+//TC=O(n)
```
Nandha2Tech, line 4: Since we are comparing two numbers together(i and i+1), for loop will iterate only for half the time.(i is incremented by i+2). So time complexity will be O(n/2).

Nandha2Tech (commented):
Good job. Kindly check time complexity for MaxMinInArray.
Since we are comparing two numbers together(i and i+1), for loop will iterate only for half the time.(i is incremented by i+2). So time complexity will be O(n/2).

Nandha2Tech:
Good job. Kindly check time complexity for MaxMinInArray.
Since we are comparing two numbers together(i and i+1), for loop will iterate only for half the time.(i is incremented by i+2). So time complexity will be O(n/2).

PR #1384, opened 2024-01-05
Array-2 complete

Dibyajyoti227:


Problem1.py
```diff
@@ -0,0 +1,20 @@
+'''
+Time complexity: O(n) because it traverses through each value in the array
+Space complexity: O(n) , not sure.
```
ishika2741, line 3: It is O(1) , output space does not count

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1427, opened 2024-02-14
Solved both problems!

pxb5080:


vyankateshmisar:
Good work Prakhar. Please include Tc and Sc in every problem.

vyankateshmisar:
Good work Prakhar. Please include Tc and Sc in every problem.

PR #1490, opened 2024-07-03
completed array-2

dhikki-study:
############448. Find All Numbers Disappeared in an Array########################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : first I coded using hashes to save al values and loop through it to find the missing value but after class learnt way to play with array itself

// Your code here along with comments explaining your approach in three sentences only
We will iterate through the array and as per the stored value we will update the index by multiplying with -1 so the index which is positive is the missing number and all negative are the numbers which are present	

        

###########289. Game of Life##########################################################
// Time Complexity : mXn
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : first I coded using another matrix where I stored net phase but post class I understood how it could be done in same matrix

// Your code here along with comments explaining your approach in three sentences only
Go through each element and check if in next phase it can be alive or dead on that basis updated cell to 3 if it moving for dead to alive and 4 in case it is alive to dead now while counting neighbor we can keep a track of old phase using 3 and 4

        

##########https://www.geeksforgeeks.org/maximum-and-minimum-in-an-array/############################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : 1st coded my way thank coded as per class which reduced O(n) to 1.5N

// Your code here along with comments explaining your approach in three sentences only
#in my way I appended extra character at end and iterated through each element comparing with 1st and last element in way that 1st element is smaller and last element is largest and at end the smallest value will be present in 0th index and max value will be present in last index
# in class it was taught to take 2 element at a time and compare themselves to check which is smallest and than we compared smaller with min and larger one with max and we did this task till end off array

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,136 @@
+############448. Find All Numbers Disappeared in an Array###############################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : first I coded using hashes to save al values and loop through it to find the missing value but after class learnt way to play with array itself
+
+
+// Your code here along with comments explaining your approach in three sentences only
+We will iterate through the array and as per the stored value we will update the index by multiplying with -1 so the index which is positive is the missing number and all negative are the numbers which are present	
+
+
+class Solution:
+    def findDisappearedNumbers(self, nums: List[int]) -> List[int]:
+        l1=[]
+        for i in nums:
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1496, opened 2024-07-04
Array 2 

PRAVEENKUMARRAJENDRAN:


gameoflife.py
```diff
@@ -0,0 +1,42 @@
+"""
+Time complexity: O(n)
```
ishika2741, line 2: you need to consider rows and columns both

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1601, opened 2025-01-17
done

rishabh-kr-jain:


Erlybird:
For the 2nd problem, you are not considering the last element of the array if array length is odd number. Please cross Check and resubmit

Erlybird:
For the 2nd problem, you are not considering the last element of the array if array length is odd number. Please cross Check and resubmit

PR #1604, opened 2025-01-18
Array 2 done

triptighanghas:


Erlybird:
In the minMax Array problem, what happens if there are odd number of elements in the array. Please cross Check the solution and resubmit

Erlybird:
In the minMax Array problem, what happens if there are odd number of elements in the array. Please cross Check the solution and resubmit

PR #1715, opened 2025-05-29
added disappered number in an array and game of life

kedarbhusari:


dalvirushikesh:
Great Work!

super30admin:
1. Correctness:
   - For `find_disappered.py`, the solution correctly identifies missing numbers in the range [1, n] where n is the length of the array. However, the global variable `nnums` is problematic as it retains state between function calls, which could lead to incorrect results if the function is called multiple times.
   - For `game_of_life.py`, the solution correctly implements Conway's Game of Life rules by checking neighbors and updating the board based on the number of live neighbors. The use of a copy board ensures that updates don't interfere with neighbor checks.

2. Time Complexity:
   - `find_disappered.py`: The time complexity is correctly stated as O(n) since it iterates through the array once to populate the dictionary and once more to check for missing numbers.
   - `game_of_life.py`: The time complexity is correctly stated as O(mn) where m and n are the dimensions of the board, as it checks each cell and its neighbors.

3. Space Complexity:
   - `find_disappered.py`: The space complexity is correctly stated as O(n) due to the dictionary storing all unique numbers.
   - `game_of_life.py`: The space complexity is correctly stated as O(mn) due to the copy of the board.

4. Code Quality:
   - `find_disappered.py`: The use of a global variable is a significant issue. The dictionary `nnums` should be initialized inside the function to avoid state retention. The function name has a typo (`find_disappered` instead of `find_disappeared`).
   - `game_of_life.py`: The code is well-structured and readable. The use of list comprehensions for copying the board is a good practice. The neighbor checks are clear and correctly implemented.

5. Efficiency:
   - `find_disappered.py`: The solution could be optimized to use O(1) space by marking the presence of numbers in the original array (e.g., by negating values at corresponding indices).
   - `game_of_life.py`: The solution is efficient for the problem, but could be optimized to use O(1) space by using special markers to denote state changes without a separate copy.

PR #1738, opened 2025-06-15
Completed

solankiram2023:


super30admin:
- The solution is correct and efficient in terms of time complexity, which is a significant improvement over the brute-force reference solution.
- The use of an auxiliary array is a straightforward approach, but the space complexity could be optimized further by using in-place marking, as hinted in the commented-out code. This would align better with the follow-up question in the problem description.
- The code is well-commented and easy to understand, which is excellent practice.
- Consider exploring the in-place marking approach to reduce space complexity, as it would make the solution more optimal and align with the follow-up challenge.

PR #1751, opened 2025-07-04
Array 2 completed

tanishkasonar22:


super30admin:
- For Problem 1: The idea of marking elements as negative is good, but the return statement should be outside the loop to check all elements. Also, `out.append(i+1)` should be `out.append(i+1)` (no assignment needed). The logic is correct but needs proper implementation.
- For Problem 2: There are syntax errors (missing colons, semicolon instead of colon). The logic for finding min and max is flawed and incomplete. The loop doesn't iterate correctly, and the comparison logic is incorrect. You need to compare all elements to find the actual min and max.
- General: Pay attention to syntax and indentation. Test your code with sample inputs to ensure correctness. Break down the problem into smaller steps and validate each step.

PR #1757, opened 2025-07-09
Array-2 Done

JPolishetty:


super30admin:
- Strengths:
  - Excellent use of in-place techniques to optimize space complexity.
  - Clear and concise logic with good comments explaining the approach.
  - Efficient time complexity for all problems.
- Areas for Improvement:
  - In Problem1, consider adding a brief comment explaining why the absolute value is used (to handle already marked indices).
  - In Problem3, the liveNeighbours method could be slightly optimized by checking bounds before accessing the array to avoid multiple checks.

super30admin:
Strengths:
- The solutions are correct and efficient, demonstrating a good understanding of the problems.
- The code is clean, well-commented, and follows best practices.
- The use of in-place techniques (Problem1 and Problem3) and pairwise comparisons (Problem2) shows advanced problem-solving skills.

Areas for Improvement:
- In Problem1, the comment could be clearer about why we are making the index negative (to mark the presence of the number).
- In Problem2, the loop condition could be simplified to `i < arr.length` and handle the odd case inside the loop for better readability.
- In Problem3, the `liveNeighbours` method could be made static since it doesn't use any instance variables.

PR #1779, opened 2025-08-14
Completed Array-2

AneerudhS27:


super30admin:
- Strengths:
  - The student demonstrates a strong understanding of efficient algorithms and in-place operations.
  - The code is clean, well-commented, and easy to follow.
- Areas for Improvement:
  - In `findDisappearedNumbers`, the comment could be more detailed to explain why the negative marking technique works.
  - In `gameOfLife`, the `dirs` vector could be made `static const` to avoid reinitialization for each instance.
  - In `findMinAndMax`, the use of `std::min` and `std::max` is good, but the student could consider using `const` for the input vector to indicate it won't be modified.

PR #1781, opened 2025-08-26
Array-2_Done

Sarthak15997:


super30admin:
- Strengths:
  - The solution is correct and follows the optimal approach.
  - The use of a helper function (`getCount`) to count live neighbors is a good practice.
- Areas for Improvement:
  - While the solution is correct, it would be beneficial to add a brief explanation of the temporary states (2 and 3) in the comments for clarity, especially for readers unfamiliar with the problem.
  - The variable names (`cnt`, `dx`, `dy`) are clear, but using more descriptive names like `count` or `delta_x`, `delta_y` could further improve readability.

PR #1796, opened 2025-09-06
Array2 Done

desaiparam:


super30admin:
- For `findDisappearedNumbers`, while the solution is correct and efficient in terms of time, it could be optimized further to reduce space complexity. For example, you could use the input array itself to mark the presence of numbers (by negating values) and then collect the missing numbers, achieving O(1) space complexity.
- For `gameOfLife`, the solution is well-implemented. One minor improvement could be to use more descriptive variable names for `nr` and `nc` (e.g., `new_row` and `new_col`) to enhance readability.
- Overall, the code is clean and follows best practices. The explanations in the comments are helpful and demonstrate a good understanding of the problems.

PR #1823, opened 2025-11-27
Completed Array-2

pranavs0133:


super30admin:
Strengths:
- Excellent use of in-place marking techniques to solve both problems without additional space.
- Clear and concise comments that explain the approach.
- Proper handling of edge cases and boundary conditions.
- Efficient time and space complexity for both problems.

Areas for Improvement:
- For Problem1, while the solution is correct, it might be worth mentioning why you're using absolute values (to handle already marked negative values).
- For Problem2, the method name `countLiveNeighbours` could be more consistent with Java naming conventions (should be `countLiveNeighbors`).
- Consider adding a brief explanation of the direction array in Problem2 for better clarity.

PR #1853, opened 2026-02-19
Complete Array-2

PrakarshKamal:


super30admin:
Strengths:
- You implemented an optimal solution with O(n) time and O(1) space complexity.
- The code is clean and well-commented, showing both an initial approach and an optimized one.
- You used the in-place marking technique correctly by negating numbers to indicate presence.

Areas for Improvement:
- In the commented O(n) space solution, there is a bug: you are adding `i+1` when the set does not contain `i`. This should be `i` instead, because `i` is already the number in the range [1, n]. For example, if n=2 and i=1 is missing, you should add 1, not 2. Similarly, if i=2 is missing, you should add 2, not 3. So please correct that if you ever use that approach.
- The file name "DisappearingNumbers.java" is slightly off from the problem name "Disappeared Numbers", but this is minor.

Note: You included code for "Game of Life" in the same submission. This might be a mistake because the problem is only about "Disappeared Numbers". In the future, make sure to submit only the code for the problem you are solving.

PR #1856, opened 2026-02-21
Completed Array-2

sarvanibaru:


super30admin:
Your solution for the "Disappeared Numbers" problem is excellent. It is efficient, both in time and space, and correctly solves the problem. The comments are helpful and explain the approach well. However, note that the problem statement was for a C++ solution, but you provided a Java solution. While the algorithm is language-agnostic, it is important to follow the problem requirements. In this case, the problem did not specify a language, but the reference solution was in C++. Since the problem is from LeetCode, which supports multiple languages, this is acceptable. But always ensure you are using the language expected by the context.

Also, you included two additional files that are not related to the problem. When submitting solutions, make sure to only include the relevant code to avoid confusion.

Overall, great job on the solution!

PR #1858, opened 2026-02-22
FAANMG | Arrays-2

sandy7907:


super30admin:
Your solution for "Disappeared Numbers" is excellent! It efficiently solves the problem with optimal time and space complexity. The approach of using the array itself to mark presence is clever and meets the constraints. However, note that you included an unrelated solution (GameOfLife) in the same file. Make sure to submit only the required code for the problem you are solving. Also, consider adding brief comments to explain the marking step for clarity. Keep up the good work!

PR #1859, opened 2026-02-22
Add implementation for leetcode problems 289, 448

rishigoswamy:


super30admin:
Strengths:
- You attempted to use an efficient cyclic sort approach, which is a good strategy for this problem.
- You recognized the need for in-place modification to achieve O(1) space complexity.

Areas for Improvement:
1. **Logic Error**: The condition in the while loop is incorrect. The correct approach for cyclic sort is to place each number at its correct index (i.e., the number `x` should be at index `x-1`). You should compare the current number with the number at its target index to decide whether to swap.
   - Correct condition: While the current number is not at its correct index (i.e., `nums[i] != nums[nums[i]-1]`), swap it with the number at its correct index. But be cautious of duplicates to avoid infinite loops.

2. **Variable Naming**: Use descriptive variable names. For example, instead of `swapIdx`, you might use `correct_index` or `target_index` to clarify that it is an index.

3. **Code Clarity**: Add comments to explain the steps. For cyclic sort, you can explain that you are trying to place each number in its correct position and then iterate to find indices that don't have the correct number.

4. **Edge Cases**: Consider cases with duplicates. The cyclic sort should avoid infinite loops by not swapping when the target index already has the correct number.

Here is a corrected version of the cyclic sort approach for your reference:
```python
def findDisappearedNumbers(self, nums: List[int]) -> List[int]:
    i = 0
    n = len(nums)
    while i < n:
        correct_index = nums[i] - 1
        if nums[i] != nums[correct_index]:
            nums[i], nums[correct_index] = nums[correct_index], nums[i]
        else:
            i += 1

    res = []
    for i in range(n):
        if nums[i] != i+1:
            res.append(i+1)
    return res
```
Note: This version uses a while loop with an index `i` that increments only when the current number is already at the correct position or when a swap might cause an infinite loop (due to duplicates). This ensures that each element is placed at its correct position at most once.

PR #1860, opened 2026-02-24
Problem solutions.

Denishmoradiya:


super30admin:
Your solution for the "Disappeared Numbers" problem is correct and efficient in terms of time complexity. The use of a HashSet allows for O(1) lookups, making the overall algorithm O(n). However, the problem's follow-up challenge asks for a solution without extra space. While your solution is acceptable for most cases, you could consider an alternative approach that uses the original array to mark the presence of numbers without using extra space. For example, you could iterate through the array and mark the index corresponding to each number as negative (if the array is mutable) to indicate that the number has been seen. Then, in a second pass, collect all indices that are still positive. This would achieve O(n) time and O(1) extra space (excluding the output list).

Strengths:
- Code is well-structured and readable.
- Comments are clear and explain the steps.
- Correctly handles the problem constraints.

Areas for Improvement:
- To meet the follow-up challenge, try to implement the in-place marking solution to reduce space complexity to O(1) extra space.
- Ensure that you only submit the solution for the problem at hand. The other files (Problem2.java and Problem3.java) are not part of this problem and should be submitted separately.

PR #1861, opened 2026-03-02
Adding all the solutions

subbu4061:


super30admin:
Your solution for the "Disappeared Numbers" problem is excellent. You have implemented an optimal solution with O(n) time and O(1) space (excluding the output). The approach of marking indices by negating values is efficient and avoids using extra data structures. The code is clear and well-commented.

However, note that the problem statement was for a C++ solution, but you provided a Java solution. While the algorithm is language-agnostic, it's important to follow the problem requirements. In this case, since the problem was presented with a C++ reference solution, you should ensure you are writing in the correct language. But since the problem is about the algorithm, and your solution is correct and efficient, this is a minor point.

For future submissions, make sure to use the language specified in the problem. Also, while your comments are helpful, avoid including multiple solutions in one submission unless asked.

PR #1862, opened 2026-03-03
Array 2 done

pranjay01:


super30admin:
Strengths:
- You implemented an optimal solution for the "Disappeared Numbers" problem with O(n) time and O(1) space.
- You correctly used the negation technique to mark the presence of numbers without extra space.

Areas for Improvement:
- The solution includes a test call at the end (print statement). In a typical submission (e.g., on LeetCode), the solution should only contain the class and method definitions without any test code. Remove such test calls to avoid runtime errors in automated judging systems.
- Although not necessary, you could add a brief comment explaining the purpose of the first loop (marking indices) and the second loop (collecting results) for clarity.

Overall, your solution for "Disappeared Numbers" is excellent and meets all requirements.

PR #1866, opened 2026-04-22
Completed Array-2

pratikb0501:


super30admin:
### Disappeared Numbers (Problem_1.py)
```

First, I need to evaluate the student's solution for the problem "Disappeared Numbers". The problem is to find all numbers in the range [1, n] that are missing from the array, where n is the length of the array.

The student's solution is written in Python. Let me look at the code:

```python
class Solution:
    def findDisappearedNumbers(self, nums):
        result = []
        n = len(nums)
        for i in range(n):
            curr = abs(nums[i])
            if nums[curr - 1] > 0:
                nums[curr - 1] *= -1
        for i in range(n):
            curr = nums[i]
            if curr > 0:
                result.append(i + 1)
            else:
                nums[i] *= -1
        return result
```

The student also provided a test case at the end, but that might not be part of the solution they intend to submit. I'll focus on the method.

Now, let's understand what the code is doing.

The approach seems to be using the array itself to mark which numbers are present. Since the numbers are in the range [1, n], we can use the indices to represent the numbers.

In the first loop, for each element in nums, it takes the absolute value (which might not be necessary if all numbers are positive, but nums can have negative values after marking). Then, it checks the value at index `curr - 1`. If that value is positive, it multiplies it by -1 to mark it as negative. This is a common technique to mark the presence of a number without using extra space.

In the second loop, it iterates through the array again. For each index i, it checks if the value is positive. If it is, that means the number (i+1) was not marked as present, so it appends i+1 to the result. Otherwise, if it's negative, it multiplies it by -1 to restore the original value? Wait, no, the problem doesn't require restoring the array, but in this case, the student is setting it back to positive in the else clause. However, the problem says "return the array of disappeared numbers", so modifying the input array might be acceptable as long as we return the correct result. But typically, we should avoid modifying the input if not necessary, but here it's allowed since we're using O(1) space.

But let's see: in the second loop, when the value is negative, it sets `nums[i] *= -1`, which restores the original positive value. So after this loop, the array should be back to its original state, except that some values might have been flipped twice? No, because in the first loop, we mark some indices negative, but in the second loop, for indices that are negative, we set them to positive. But for indices that are positive in the second loop, we don't change them. However, the indices that were marked negative in the first loop are set to positive in the second loop, so the array is restored. But is that correct?

Actually, in the first loop, we are marking the index (curr - 1) as negative. But when we iterate in the second loop, we are looking at each index i. If the value at i is positive, it means that index was never marked negative, so number i+1 is missing. If it's negative, we set it back to positive.

But there's a potential issue: when we mark an index in the first loop, we might be marking an index that has already been marked? But since we are using absolute value, it should be fine.

Now, let's test with the example: nums = [4,3,2,7,8,2,3,1]

n = 8

First loop:

i=0: nums[0] = 4, curr = abs(4) = 4, index = 4-1=3, nums[3] = 7 >0, so set nums[3] = -7

Array: [4,3,2,-7,8,2,3,1]

i=1: nums[1] = 3, curr=3, index=2, nums[2]=2>0, set to -2

Array: [4,3,-2,-7,8,2,3,1]

i=2: nums[2] = -2, but we take abs, so curr=2, index=1, nums[1]=3>0, set to -3

Array: [4,-3,-2,-7,8,2,3,1]

i=3: nums[3] = -7, abs=7, index=6, nums[6]=3>0, set to -3

Array: [4,-3,-2,-7,8,2,-3,1]

i=4: nums[4]=8, curr=8, index=7, nums[7]=1>0, set to -1

Array: [4,-3,-2,-7,8,2,-3,-1]

i=5: nums[5]=2, curr=2, index=1, nums[1]=-3<0, so no change? The condition is if nums[curr-1] >0, so since -3 is not >0, we skip.

Array unchanged.

i=6: nums[6]=-3, abs=3, index=2, nums[2]=-2<0, skip.

i=7: nums[7]=-1, abs=1, index=0, nums[0]=4>0, set to -4

Array: [-4,-3,-2,-7,8,2,-3,-1]

Now, after first loop, the array is: [-4,-3,-2,-7,8,2,-3,-1]

Now second loop:

i=0: nums[0]=-4 <0, so we set it to positive: nums[0]=4, and do nothing for result? Only if positive we add to result.

So for i=0, since it's negative, we set to positive but don't add to result.

i=1: nums[1]=-3<0, set to positive:3

i=2: nums[2]=-2<0, set to positive:2

i=3: nums[3]=-7<0, set to positive:7

i=4: nums[4]=8>0, so we add i+1=5 to result.

i=5: nums[5]=2>0, add i+1=6 to result.

i=6: nums[6]=-3<0, set to positive:3

i=7: nums[7]=-1<0, set to positive:1

So result is [5,6], which is correct.

Also, the array is restored to [4,3,2,7,8,2,3,1], which is the original.

So the solution seems correct.

Now, about time complexity: the first loop is O(n), second loop is O(n), so overall O(n). This is better than the reference solution which is O(n^2) and times out.

Space complexity: O(1) extra space, since we are using the input array and the result list is not counted as extra space. So it meets the requirement.

Code quality: The code is fairly readable, but there are some issues.

- The variable names are not very descriptive. For example, `curr` could be better named as `num` or `val`.

- The logic might be a bit tricky to understand at first glance. Adding comments would help.

- In the second loop, the else clause is restoring the array, which is not strictly necessary for the problem, but it's a good practice if we want to preserve the input. However, the problem doesn't specify whether we can modify the input or not. Since it's allowed to use O(1) space, modifying the input is acceptable, but restoring it might be an extra step that isn't needed. Actually, in this case, if we don't restore, the input array would be modified, which might not be desired. But the problem doesn't say anything about modifying the input, so it's fine either way. However, in competitive programming, it's often acceptable to modify the input as long as the output is correct.

- There's a potential issue: when we do `nums[curr - 1] *= -1`, we assume that the value at that index is positive. But if it's already negative, we don't change it. That's correct because we are marking presence, so if it's already marked, we don't need to mark again.

- Also, in the first loop, we use `abs(nums[i])` to handle the case where the value might have been negated in previous steps. This is correct.

Overall, the solution is efficient and correct.

But let's check the test case provided at the bottom: the student has written a print statement that might not be appropriate in a submission. But since this is probably just for testing, it might be okay. In a real Leetcode submission, they should remove the test code.

Now, comparing to the reference solution: the reference solution is a brute force O(n^2) that checks for each number from 1 to n if it exists in the array. It's inefficient and will time out for large n. The student's solution is much better.

So, in terms of correctness, time complexity, space complexity, it's good.

Code quality could be improved with better variable names and comments.

But

VERDICT: NEEDS_IMPROVEMENT

---

### max and min (Problem_2.py)
Your approach to process elements in pairs is good, as it reduces the number of comparisons. However, there are several issues to address:

- For an even-length array, you initialize both min and max to arr[0]. This is correct only if you then compare with the first pair? Actually, in the reference solution, for even length, it compares the first two elements to set min and max. Then it starts from index 2. Your code for even length does not do any comparison for the first two elements; it just sets both to arr[0]. Then you start from index 2. This means you never compare the first element with anything else? So if the first element is the maximum, but you set maxi to arr[0] and then never update it if all other elements are smaller, that would be wrong. Similarly for minimum.

- For an odd-length array, you are comparing the first two elements (index0 and index1) to set min and max. But what if the array has only one element? Then n=1, and you try to access arr[1] which is out of bounds. You should handle the case when n==1 separately.

- The reference solution handles the odd-length case by setting both min and max to the first element, and then starting from index1 (which is the second element). Then it processes pairs from there. For even length, it processes the first pair and then starts from index2.

- Your while loop condition is "i < n-1", which is correct for processing pairs (since you access i and i+1). But for an odd-length array, if you start at i=2 (after initializing with the first two), then you will process pairs from index2 to n-2? For example, if n=5, you start i=2, then compare index2 and3, then i becomes4. Then i=4 is not less than n-1=4? So the loop stops. The last element (index4) is never compared. So you miss the last element.

- You should follow the reference solution more closely. Initialize min and max appropriately based on whether n is even or odd. For even: compare first two. For odd: set min and max to first element. Then set i to the next index (2 for even, 1 for odd). Then process pairs.

- Also, you should use proper indentation in Python. The while loop should be indented to the same level as the if-else block.

Here is a corrected version of your code in Python:

```python
def find_min_max(arr):
    n = len(arr)
    if n == 0:
        return (None, None)  # or handle empty array
    if n % 2 == 0:
        if arr[0] < arr[1]:
            mini, maxi = arr[0], arr[1]
        else:
            mini, maxi = arr[1], arr[0]
        i = 2
    else:
        mini, maxi = arr[0], arr[0]
        i = 1

    while i < n - 1:
        if arr[i] < arr[i+1]:
            mini = min(mini, arr[i])
            maxi = max(maxi, arr[i+1])
        else:
            mini = min(mini, arr[i+1])
            maxi = max(maxi, arr[i])
        i += 2

    return (mini, maxi)
```

This code should work correctly. Note that for an odd-length array, the last element is not processed in the loop? Actually, if n is odd, we start at i=1 (which is the second element) and then process pairs until i < n-1. Since n is odd, the last index is even (0-indexed). For example, n=5: indices 0,1,2,3,4. We set i=1. Then we process pairs: i=1: compare index1 and2; then i=3: compare index3 and4. Then i=5 which is not <4. So it works.

VERDICT: NEEDS_IMPROVEMENT

---

### Life Game (Problem_3.py)
It appears that you did not submit a solution for this problem. To successfully complete this task, you need to implement the gameOfLife function that updates the board in-place according to the rules of Conway's Game of Life. 

Here are some key points to consider when implementing your solution:
1. The problem requires updating the board in-place without returning anything.
2. Since updates must be simultaneous, you cannot update cells directly using the new states of neighbors. You need a way to mark changes without affecting the current state.
3. A common approach is to use intermediate states to mark cells that will change. For example:
   - Use 2 to mark a live cell that will become dead.
   - Use 3 to mark a dead cell that will become live.
4. Then, in a second pass, convert all intermediate states to their final values (0 or 1).
5. You'll need to check all 8 neighbors for each cell, making sure to handle boundaries correctly.

I encourage you to try implementing the solution and resubmit. If you need help, refer to the reference solution provided.

VERDICT: NEEDS_IMPROVEMENT

